PR #1, opened 2019-07-10
Problem 2: Please review

hrshah5:


Problem2.java
```diff
@@ -0,0 +1,18 @@
+// Time Complexity : O(n)
+// Space Complexity :O(1)
+// Did this code successfully run on Leetcode : No, passed 39 cases. I didn't took account of values which which can be more than 2 distance and still can be more in outcome. Ex: [2,1,1,2]
+// Any problem you faced while coding this :
+class Solution {
```
super30admin: This approach wont work here.
Because you need to check both the elements at distance n+2 as well as n+3;

Problem2(geeksforgeeks).java
```diff
@@ -0,0 +1,29 @@
...
+        }
+        int maxMoney = 0;
+        for(int i = 2; i<nums.length; i++){
+            maxMoney = Math.max(nums[i] + value0, value1);
+            value0 = value1;
+            value1 = maxMoney;
+        }
+        return maxMoney;
+        
+
```
super30admin: Practice this Solution yourself and try solving similar kind of problem.

PR #6, opened 2019-07-11
brute force solution needs alot of improvement

sehaj-grover:


super30admin:
Well tried but these Solutions wont work in these problems.
You need to think a lot more.
GoodLuck 👍

PR #7, opened 2019-07-11
Completed Day 7 - Please Review

bharatreddypothireddy:
Not a correct solution . Did not execute for all test cases on leet code. Confused as to how to approach solving it.

### DIFF
--- a/CoinChange.java
+++ b/CoinChange.java
@@ -1,7 +1,7 @@
-// Time Complexity :O(n*n) --> beacuse of for loop
-// Space Complexity :O(n)  ---> because of Hashmap
-// Did this code successfully run on Leetcode : No
-// Any problem you faced while coding this : Didn'' understand how to approach for different test cases
+// Time Complexity :O(M*N) --> M is number of coins and N is amount
+// Space Complexity :O(M*N)  ---> because of 2d array
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : No.
 
 
 // Your code here along with comments explaining your approach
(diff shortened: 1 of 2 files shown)


HouseRobber.java
```diff
@@ -0,0 +1,20 @@
+// Time Complexity :O(n) --> beacuse of for loop
+// Space Complexity :O(1)  ---> because of Hashmap
+// Did this code successfully run on Leetcode : No
+// Any problem you faced while coding this : Didn'' understand how to approach for different test cases
+
+
+// Your code here along with comments explaining your approach
+
+class Solution {
+    public int rob(int[] nums) {
```
super30admin: Its not the case of adding even and odd.
You need to where to go (n+2 or n+3) whichever is maximum out of these two.

CoinChange.java
```diff
@@ -0,0 +1,37 @@
+// Time Complexity :O(n*n) --> beacuse of for loop
+// Space Complexity :O(n)  ---> because of Hashmap
+// Did this code successfully run on Leetcode : No
+// Any problem you faced while coding this : Didn'' understand how to approach for different test cases
+
+
```
super30admin: This Solution wont work.
Try once again using Knapsack approcah.

HouseRobber.java
bharatreddypothireddy: Completed using dynamic programming approach by finding the maximum amount a robber can rob on reaching a particular house at index 'i' and adding them to a dp array . Finally return the last value of the array.

CoinChange.java
bharatreddypothireddy: Completed using tabulation method approach taught by Jaspinder.

jaspindersingh83:
Good

PR #10, opened 2019-07-11
Please review

Shazmaan:
Missing runtime but will add that soon.

### DIFF
--- a/Problem2.cpp
+++ b/Problem2.cpp
@@ -15,32 +15,54 @@ using namespace std;
 
 class Solution{
 public:
-    int prob(vector<int>& nums){
-        int index = 2; int sum=nums[0]; vector<int> allNum;
-        while(index<nums.size()){
-            if(index+2<nums.size()){
-                if(nums[index+1]>nums[index]){
-                    if(nums[index+1]<nums[index+2]){
-                        sum+=nums[index]; allNum.push_back(nums[index]);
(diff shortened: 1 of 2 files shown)


Problem2.cpp
```diff
@@ -15,32 +15,54 @@ using namespace std;
...
+//                    sum+=nums[index];
+//                }else if(index-2>=0 && find(allNum.begin(),allNum.end(),nums[index-2])!=allNum.end()){
+//                        sum+=nums[index];
+//                }
+////                }
+//            }
+//            cout<<nums[index]<<" "<<sum<<endl;
+//            index+=2;
+//        }
+//        return sum;
```
super30admin: This seems be a lot better version as compared to the first solution.

Problem2.cpp
```diff
@@ -15,32 +15,54 @@ using namespace std;
...
+        // Fill remaining positions
+        for (int i=2; i<size; i++)
+        {
+            maxVal = max(nums.at(i)+ind1, ind2);
+            ind1 = ind2;
+            ind2 = maxVal;
         }
-        return sum;
+
+        return maxVal;
```
super30admin: This seems be a lot better version as compared to the first solution.

jaspindersingh83:
The DP solution of problem 1 is missing still. Very important. Please submit

jaspindersingh83:
Good

PR #24, opened 2019-08-29
Completed: Kindly review the DP Problems 

DeekanshaTandon:


robber.java
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        if(nums.length ==1){
+            return nums[0];
+        }
+
+        int[] cache = new int[nums.length+1];
+        //Base cases are declared
+        cache[0]=nums[0];
+        cache[1]=Math.max(nums[0],nums[1]);
```
kaurramanpreet: great way of starting and not using extra space for two more elements!!

PR #27, opened 2019-08-30
DP-1 done

pradhan1234:
Please review @kaurramanpreet

### DIFF
--- a/HouseRobber.java
+++ b/HouseRobber.java
@@ -1,0 +1,50 @@
+/**
+ * You are a professional robber planning to rob houses along a street. Each
+ * house has a certain amount of money stashed, the only constraint stopping you
+ * from robbing each of them is that adjacent houses have security system
+ * connected and it will automatically contact the police if two adjacent houses
+ * were broken into on the same night.
+ * 
+ * Space Complexity: O(n) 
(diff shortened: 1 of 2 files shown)


HouseRobber.java
```diff
@@ -0,0 +1,50 @@
...
+        if (len == 2) {
+            return Math.max(nums[0], nums[1]);
+        }
+        // using dynamic programming to store solutions to overlapping subproblems.
+        int[] dp = new int[len];
+        dp[0] = nums[0];
+        dp[1] = Math.max(nums[0], nums[1]);
+        for (int i = 2; i < len; i++) {
+            // should we choose nums[i] depends on whether nums[i] + dp[i-2] 
+            // is optimal then dp[i-1]
```
kaurramanpreet: can you elaborate your doubt
i didnt get it!!

pradhan1234: Not doubt, it was a comment to jot down my reasoning let me rephrase it :)

pradhan1234: "To Be or Not to be", 
that's the decision that drives dp, 

so will nums[i] be included (robbed) depends if nums[i] + dp[i-2] is more than dp[i], in which we skip robbing current house

PR #30, opened 2019-08-31
Completed

PATIL1695:
Please let me know why we need this intialization -   dp[0], dp[1] = nums[0], max(nums[0], nums[1])

### DIFF
--- a/HouseRobbert.py
+++ b/HouseRobbert.py
@@ -1,0 +1,19 @@
+"""
+The approach here is to keep on computing the the current index + the index - 2 values sum and the 
+index - 1 sums and compare them for max values which ever is maximum store that in the dp array.
+
+"""
+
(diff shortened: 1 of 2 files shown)


HouseRobbert.py
```diff
@@ -0,0 +1,19 @@
...
+
+"""
+
+def rob(self, nums):
+        if len(nums)==0:
+            return 0
+        if len(nums)==1:
+            return nums[0]
+        dp = [0]*len(nums) 
+        dp[0], dp[1] = nums[0], max(nums[0], nums[1])
```
kaurramanpreet: because in the loop we are trying to go to location which is 2 indices before that index so we need to take care of the fact that we dont go out of the bounds and so we initialized our oth and 1st indices before going into the loop

PR #32, opened 2019-10-16
practice

pkbaharani:


Problem-198
```diff
@@ -0,0 +1,24 @@
+/*
+Time Complexity: O(n)
+Did not run in leetcode: 39 / 69 test cases passed.
+Approach: The only solution i could think of was that once check odd positions and once check even positions.
+I understand that this is totaly incorrect, but i could not come up with an algorithm to accomodate best houses.
+
```
sehaj-grover: try finding the max you can get from any given 2 neighbours at a time and keep moving forward

Problem-322
```diff
@@ -0,0 +1,55 @@
...
+Time Complexity- > O(n);
+
+46 out of 182 test cases passed.
+
+Approach: sorted the given array.
+
+Always pick the biggest denomination. and keep on adding the coins 
+add to the result.
+
+
```
sehaj-grover: We might skip some possible cases if we do that
for 11 using 2,5 if I add 5+5 i can't make 11 using 2 So we need to fix this condition

PR #35, opened 2019-10-17
Prob-1 & 2 solutions added

akhileshborgaonkar:


akhileshborgaonkar:
Prob-1 solution is not optimal. I will work on it after session.

sehaj-grover:
very good Akhilesh 
Do submit it now after reviewing the session

akhileshborgaonkar:
Optimized solutions added.

sehaj-grover:
great!!

PR #39, opened 2019-10-17
Completed 19 & 20

sindhu819:
I got error for both the problems while solving using recursive method  - Maximum Recursive depth exceeded error.

### DIFF
--- a/Problem-19.py
+++ b/Problem-19.py
@@ -1,0 +1,78 @@
+# leet code -322
+# Dynamic Programming
+# passed all test cases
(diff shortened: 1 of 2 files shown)


sindhu819:
Not yet. Still figuring it out.

sehaj-grover:
Which problem are you unable to understand
Also which technique?

sindhu819:
I resolved my error. I understood where I went wrong in recursive case.

sehaj-grover:
great!

PR #48, opened 2019-11-28
Completed DP-1

SweetySojrani:


Problem1.java
```diff
@@ -0,0 +1,51 @@
+// Brute Force Solution
+// Time Complexity - O(S^n)
+// Space Complexity - O(n^2)
+// This Solution did not work Successfully on Leetcode because it is not an optimized solution and the time limit exceeded.
+
```
neha79091: Think again for the time and space complexities.

PR #56, opened 2019-12-12
Update houserob.java

sakshi1705:


houserob.java
```diff
@@ -0,0 +1,19 @@
+# Time Complexity : O(n)
+# Space Complexity : O(1)
+
```
neha79091: Space complexity is O(n)

PR #59, opened 2020-01-02
DP-1: Problem2 - Robber

BhushanVasisht:


super30admin (commented):
` else if(nums.length == 1)
            return nums[0];
        else if(nums.length == 2)
            return Math.max(nums[0], nums[1]);` This is an overkill. Good Solution though

PR #74, opened 2020-01-07
DP-1 , Problem-1

nehabindle:
Please review Solution for DP-1, Problem -1 
Ran successfully on leetcode as well


super30admin:
As discussed in class House robber can be done without extra space as well

PR #89, opened 2020-02-13
DP-1 Solutions

saiswaroopreddy:


HouseRob.java
```diff
@@ -0,0 +1,23 @@
...
+//LeetCode : Yes
+
+//Since its the maximum amount that is to be found, just initialise two variable based on the decision to select the house or not. 
+//A temporary variable to store the previous value. Maximum of the  final sel and dsel values will the maximum amount that can be robbed. 
+
+
+class Solution {
+    public int rob(int[] nums) {
+        //base
+        if(nums == null || nums.length ==0)return -1;
```
Lohith-Aswathappa: the base case should return 0 and not -1

PR #106, opened 2020-03-23
Adding code for DP-1  : Please check

RutujaKaushike:


CoinChange1.java
```diff
@@ -0,0 +1,49 @@
...
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : No
+
+import java.util.Arrays;
+
+public class CoinChange1 {
+    public static int coinChange(int[] coins, int amount) {
+        // Sorting the coins array
+        Arrays.sort(coins);
```
devashishnyati: Do we actually need to sort the array here? Think!

devashishnyati:
Good job Rutuja!
Think about sorting the array and in the class if it's still a doubt.

PR #107, opened 2020-03-24
Solution

radhika-github:


super30admin (commented):
Good Job. 

- Try House robber brute force recursive solution

- Try House Robber without extra space. In `int res = Math.max(rob(nums,dp,i-2)+nums[i], rob(nums,dp, i-1));` this code we are effective using i-2 and i-1 results for each i. Just use two variables instead of dp array to store in space optimal solution
-

super30admin:
Good Job.

Try House robber brute force recursive solution

Try House Robber without extra space. In int res = Math.max(rob(nums,dp,i-2)+nums[i], rob(nums,dp, i-1)); this code we are effective using i-2 and i-1 results for each i. Just use two variables instead of dp array to store in space optimal solution

PR #113, opened 2020-03-24
DS-1 problem-2 cmpleted

schakrad:


devashishnyati:
Good job Sindhu!
Please submit the other problem also.
Also, submit the brute force solutions.

PR #114, opened 2020-03-24
DP 1

vaidyapawan:
198. House Robber
322. Coin Change

### DIFF
--- a/198. House Robber
+++ b/198. House Robber
@@ -1,0 +1,28 @@
+Time complexity: May be O(n)?
+Space complexity: O(1)
+Did this code submitted successfully on leetcode: Yes
+
+class Solution {
+    public int rob(int[] nums) {
+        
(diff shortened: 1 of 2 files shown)


322. Coin Change
```diff
@@ -0,0 +1,26 @@
+Time complexity: ? (Never tried a DP problem before, so quite confused about the complexity)
```
devashishnyati: Its going to be O(mn) and space will also be O(mn). m is number of coins and n is amount.

198. House Robber
```diff
@@ -0,0 +1,28 @@
+Time complexity: May be O(n)?
```
devashishnyati: You are right

devashishnyati:
Good job Pawan!
Please verify the time and space complexities. 
Also, submit the brute force solutions.

PR #115, opened 2020-03-24
Completed DP1

AroopN:


devashishnyati:
Good job Aroop!
Please submit the brute force solutions also.

PR #117, opened 2020-03-24
DP1 - Please review

megrao:


devashishnyati:
Good job Megha!
Please submit the brute force solution also.

PR #120, opened 2020-03-27
DP-1 is complete. Please review and share feedback.

techielove92:
DP-1 is complete. Please review and share feedback.


Problem-1.py
```diff
@@ -0,0 +1,41 @@
...
+Input: coins = [1, 2, 5], amount = 11
+Output: 3
+Explanation: 11 = 5 + 5 + 1
+Example 2:
+
+Input: coins = [2], amount = 3
+Output: -1
+Note:
+You may assume that you have an infinite number of each kind of coin.
+// Time Complexity :  O(n^2) , not sure
```
devashishnyati: It should be O(m*n). Both time and space.

devashishnyati:
Good job Lovenish!
Also, submit the brute force solution.

PR #142, opened 2020-05-06
Please review both the problems. 

albertcoder:


super30admin:
Good Job Albert. The brute force solution for both problems are missing

super30admin:
Good Job Albert. The brute force solution for both problems are missing

albertcoder:
Hi Jaspinder, I have submitted the brute force solutions as well. Please review.

PR #146, opened 2020-05-07
DP-1, please review

Siyami:


super30admin:
Brute force of both problem missing. DP solutions are correct

PR #170, opened 2020-06-04
Add files via upload

trivedipratul:


PA_19.py
```diff
@@ -0,0 +1,60 @@
+#Recursive Solution
+#Time Complexity - Exponential
```
Lohith-Aswathappa, line 2: Be specific about time complexities, for now, 2 power n, since you are making 2 choices at every point

PA_20.py
```diff
@@ -0,0 +1,86 @@
+#LC 198 - House Robber
+#Recursive Solution
+#Time Complexity - Exponential (what will be the exact? 2^n?)
```
Lohith-Aswathappa, line 3: 2 power n, since you are making 2 choices at every point

PA_20.py
```diff
@@ -0,0 +1,86 @@
+#LC 198 - House Robber
+#Recursive Solution
+#Time Complexity - Exponential (what will be the exact? 2^n?)
+#Space Complexity - O(n) Again, doubt. Could you clear it out?
```
Lohith-Aswathappa, line 4: O(N) should be correct, recursion is called on all the element either in the chosen case or not chosen case
The recursive stack will have address of at most N/2 which is basically O(N)

Lohith-Aswathappa:
Good job, Priyansh!
Review the inline comments.

PR #194, opened 2020-06-08
DP-1 (Darshan-Review)

darshan777:


coinchange.java
```diff
@@ -1,3 +1,10 @@
+/**
+Time complexity : O(n*m) n= no of elements in coins m = amount
+Space complexity : O(n) n = new dp array of n elements
```
AbhiniveshP, line 3: Space should also be O(n * m) in your approach.

AbhiniveshP:
Nice work on both the questions !!!
Try reducing space by 1-Dimension for both the questions.

PR #210, opened 2020-07-16
Completed both problem. Kindly review it.

sneharnair:


samarthjj:
Great submission Sneha, props on including your approach that you tried before class as well!
Going forward try to use the 3-4 points approach to explain in English rather than pseudocode.

samarthjj:
Great submission Sneha, props on including your approach that you tried before class as well!
Going forward try to use the 3-4 points approach to explain in English rather than pseudocode.

PR #212, opened 2020-07-16
Please review!

arrpitsharrma:


HouseRobber.java
```diff
@@ -1,3 +1,17 @@
+
+// Time Complexity : O(n), n = number of houses
+// Space Complexity : O(n)
```
Sumesh-H, line 3: Are we using any auxiliary DS?

Sumesh-H (commented):
Good work, Arpit!

Sumesh-H:
Good work, Arpit!

HouseRobber.java
arrpitsharrma: Yes, dp[][] is auxiliary space, I think. No?

Sumesh-H: Correct. Try solving it in O(1).

arrpitsharrma: O(1) solution is is the same file below it. Kindly check and confirm.

PR #213, opened 2020-07-17
Completed DP1

alekhya-12:
- Please review. Thanks

### DIFF
--- a/CoinChange.java
+++ b/CoinChange.java
@@ -1,0 +1,44 @@
+/**
+Problem: Find the min number of coins needed to make a given amount
+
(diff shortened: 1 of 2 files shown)


samarthjj:
Good work, Alekhya. Please also include the **Recursive/Greedy** approaches as well!
Going forward try to use the comments to explain your approach in English rather than pseudocode.

samarthjj:
Good work, Alekhya. Please also include the **Recursive/Greedy** approaches as well!
Going forward try to use the comments to explain your approach in English rather than pseudocode.

PR #297, opened 2020-09-24
DP-1 solved

sujaygijre:


Lohith-Aswathappa (commented):
Good work, Sujay!
You have updated same code for both problems, check and submit

Lohith-Aswathappa:
Good work, Sujay!
You have updated same code for both problems, check and submit

PR #316, opened 2020-10-21
DP -1 - Completed- Please review

GayatriHungund81194:


coinChange.py
```diff
@@ -0,0 +1,25 @@
+# As taught in class, the  solution here will be to design a grid and memoise the result to get solution
+# Time Complexity: O(n^2)
+# Space Complexity: O(sizee of matrix)
```
samarthjj, line 3: How n^2? are the (two entities of the nested loop)/(two dimensions of the matrix) always the same?
Both time & space here are O(m * n), where m is the amount and n the number of coins.

samarthjj:
Good work Gayatri, Please review the inline comments.
Please always also submit the recursive solution for all DP problems.

samarthjj:
Good work Gayatri, Please review the inline comments.
Please always also submit the recursive solution for all DP problems.

PR #336, opened 2020-10-26
Solution to DP-1 problems

sumitjanawlekar:


coin_change.java
```diff
@@ -0,0 +1,45 @@
+// Time Complexity : O(n+m), where m is the size of coins array and n is the amount (as we are iterating over the entire grid to fill out)
```
kannan-mrugapphan, line 1: shouldn't the time be O(mn)

house_robber.java
```diff
@@ -0,0 +1,30 @@
+// Time Complexity : O(n), where n is the length of the nums array
+// Space Complexity :O(1), as we are just using 3 constant variables
+// Did this code successfully run on Leetcode : yes
+// Any problem you faced while coding this : did not understand to do this problen with single dp array (and "not" dp[nums.length][2])
```
kannan-mrugapphan, line 4: Using single dimensional dp array is exactly the same as this except for the fact that only the 2 most recent indices (i - 1, i - 2) are only used

kannan-mrugapphan (commented):
Good Job. Please try the recursive solution as well.

kannan-mrugapphan:
Good Job. Please try the recursive solution as well.

PR #371, opened 2020-12-30
Completed House Robber

shamali21:


ramya-sruthi-d (commented):
Good, but the sc is O(n) cause of using an array.

ramya-sruthi-d:
Good, but the sc is O(n) cause of using an array.

PR #402, opened 2021-01-05
DP-1

Sowmya246:


robberHouse.py
```diff
@@ -0,0 +1,25 @@
+// Time Complexity : O(n)
+// Space Complexity : O(1)
```
Lohith-Aswathappa, line 2: you are using an extra space of Nx3. your space complexity becomes O(N)

coinsChange.py
```diff
@@ -0,0 +1,18 @@
+// Time Complexity : O(s*n), s is amount and n is denomination
+// Space Complexity : O(n)
```
Lohith-Aswathappa, line 2: space would also be the same as the time complexity

Lohith-Aswathappa (commented):
Good work, Sowmya!
Check the inline comments

Lohith-Aswathappa:
Good work, Sowmya!
Check the inline comments

PR #424, opened 2021-01-27
DP-1 solution-Please Review

kapish5467:


keshav-Gupta-IS (commented):
Good work, Kapish!! You will get used to recognizing the DP questions with practice. The key is to start with the brute force solution then only we are able to find the sub problems.

keshav-Gupta-IS:
Good work, Kapish!! You will get used to recognizing the DP questions with practice. The key is to start with the brute force solution then only we are able to find the sub problems.

PR #459, opened 2021-02-25
Problem 1 completed

svyas19:


problem1.py
```diff
@@ -0,0 +1,35 @@
+import numpy as np
```
nsibal, line 1: You may or may not be allowed to use `numpy` in an actual interview. Try solving it without using a 3rd party library.

nsibal (commented):
Looks good! Read inline comments and submit the other problem as well.

nsibal:
Looks good! Read inline comments and submit the other problem as well.

PR #526, opened 2021-05-05
DP - 1

Shrinivasmiskin:
Completed both the problems


maithrreye:
Good job Shrinivas! Please add TC/SC and comments in the code.

maithrreye:
Good job Shrinivas! Please add TC/SC and comments in the code.

PR #610, opened 2021-06-30
DP-1 Solution please review

broy2808:


Stnagam:
Add Lc link, explain with comments

Stnagam:
Add Lc link, explain with comments

PR #656, opened 2021-08-13
Completed-DP-1- Please Review

prg01:


Rooppesh:
Good job with both the problems!

Rooppesh:
Good job with both the problems!

PR #706, opened 2021-09-25
Completed DP-1, Please review

Rutuja-haridas1996:


Rooppesh:
Good Job! Do add the TC, SC and approach used!

Rooppesh:
Good Job! Do add the TC, SC and approach used!

PR #784, opened 2021-12-24
please check

ramakant987:


sonalgupta276 (commented):
Good job! Mention TC and SC as well. Also write recursive solution

sonalgupta276:
Good job! Mention TC and SC as well. Also write recursive solution

PR #794, opened 2021-12-30
DP-1 attempted, please review

Mohit1215:
Finished house robber problem but coin change attempted and could not reach a final optimal solution. Attempted it with a greedy approach which fails a few test cases.

### DIFF
--- a/CoinChange.java
+++ b/CoinChange.java
@@ -90,3 +90,43 @@ class CoinChange
 
     }
 }
+
+// correct dp solution after discussed in class
+/*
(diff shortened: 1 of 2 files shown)


sanjayvarma199 (commented):
Good Job!! For Coin Change while specifying SC you can ignore lower order terms and specify as O(m*n) . For House Robber you can try using DP approach.

sanjayvarma199:
Good Job!! For Coin Change while specifying SC you can ignore lower order terms and specify as O(m*n) . For House Robber you can try using DP approach.

PR #798, opened 2021-12-30
DP1 assignment

psp65:


sonalgupta276 (commented):
Good job! Do recursive approach as well. Mention TC and SC in Big O Notation

sonalgupta276:
Good job! Do recursive approach as well. Mention TC and SC in Big O Notation

PR #975, opened 2022-06-01
Please Review

YashTike:


Sumesh-H (commented):
Good Job! Try using long or double data type for the coin change problem. I guess the test case is going beyond Integer.MAX_Value.

Sumesh-H:
Good Job! Try using long or double data type for the coin change problem. I guess the test case is going beyond Integer.MAX_Value.

YashTike:
> Good Job! Try using long or double data type for the coin change problem. I guess the test case is going beyond Integer.MAX_Value.

I don't know if that is the problem, as the code did not run even for the initial test

Sumesh-H:
> > Good Job! Try using long or double data type for the coin change problem. I guess the test case is going beyond Integer.MAX_Value.
> 
> I don't know if that is the problem, as the code did not run even for the initial test

I see the logic needs to be modified. Try again with Tabulation, connect me on slack if you're still facing trouble.

PR #1014, opened 2022-06-06
DP-1

nishanth-sayana:
I am having a doubt in Space Complexity. Please let me know about their space complexities


niharikagupta2021 (commented):
Good Job! Since you are using a 1d array for house robber problem, the sc will be O(n) and tc will be O(n) as well

niharikagupta2021:
Good Job! Since you are using a 1d array for house robber problem, the sc will be O(n) and tc will be O(n) as well

PR #1047, opened 2022-07-04
Moving LC repos to new one, review not needed

as-DSAPrep:


LC198_HouseRobber.py
```diff
@@ -0,0 +1,71 @@
...
+
+Exhaustive approach - 
+we consider choosing or not choosing each house
+
+[6,9,8,1,1,5]
+
+In this case the decision making criteria would be the current array (houses left to rob) to maximize the amount by robbing from them
+
+TC = O(n)
+SC = O(n)
```
daredevilteja, line 15: SC is O(1) for the optimized solution which you have mentioned

daredevilteja (changes requested):
Please mention TC and SC for coinchange problem

daredevilteja:
Good Work! Keep it up

PR #1066, opened 2022-07-08

rsp2210:


house_robber.py
```diff
@@ -0,0 +1,11 @@
+# Time Complexity: o(n)
+# Space Complexity: o(n)
```
daredevilteja, line 2: Here SC is O(1) because no extra space is used

daredevilteja (commented):
Good Work! Keep it up

daredevilteja:
Good Work! Keep it up

PR #1077, opened 2022-07-29
Sprint DP-1

Erv16:


Problem2.java
```diff
@@ -0,0 +1,22 @@
+// Time Complexity : O(2^n)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : No, encountered TLE
```
Sumesh-H, line 3: Try submitting the optimized solution.

Sumesh-H (commented):
Good Job! Try a different approach for the 2nd problem.

Sumesh-H:
Good Job! Try a different approach for the 2nd problem.

Problem2.java
Erv16: Submitted the optimized solution

PR #1100, opened 2022-08-07
Completed DP-1

SaitejaGarlapati:
Sorry for the late submission!


House_robber.py
```diff
@@ -0,0 +1,49 @@
+# Time Complexity : O(n)
+# Space Complexity : O(1)
```
ishika2741, line 2: It will be O(n)

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1120, opened 2022-08-30
Completed DP-1

vigviswa:
I am still looking to optimize Problem 2 further. I'll add it along when done.

Please let me know if my TC and SC are correct for the DP. If not, please kindly explain

### DIFF
--- a/houseRobber.py
+++ b/houseRobber.py
@@ -1,0 +1,46 @@
+# Time complexity: O(2^N) Exponential
+# Space complexity: O(1)
+
+# THe decision making varibale is the index of the house. We design the base case to return value if pointer reaches the end
+
+
+class Solution:
+    def rob(self, nums: List[int]) -> int:
+        return self.helper(nums, 0, 0)
+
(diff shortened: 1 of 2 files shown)


houseRobber.py
```diff
@@ -0,0 +1,46 @@
+# Time complexity: O(2^N) Exponential
+# Space complexity: O(1)
```
Sumesh-H, line 2: Recursive stack?

Sumesh-H:
Good Job with both problems. Reconsider space complexity for both problems, as we consider recursive stack as extra space.

Sumesh-H:
Good Job with both problems. Reconsider space complexity for both problems, as we consider recursive stack as extra space.

houseRobber.py
vigviswa: Yes, this is for the recursive stack. So essentially, the space complexity would be the space to construct the stack. Would it be O((2^n) - 1), as in total number of nodes of the Binary tree I construct, by choosing/not choosing?

PR #1130, opened 2022-09-14
Completed both problems for DP-1

ysoni0303:


pvk3081 (commented):
Good Job. Kindly include the brute force too

pvk3081:
Good Job. Kindly include the brute force too

PR #1393, opened 2023-08-15
code added

fastakash:


Nandha2Tech (commented):
Great work! 
Since we are using dp array SC would be O(n) for both the problems

Nandha2Tech:
Great work! 
Since we are using dp array SC would be O(n) for both the problems

PR #1458, opened 2023-11-22
DP-1

aditird:


Nandha2Tech (commented):
Great Work! Kindly include TC and SC. kindly check house robber problem.
This is how dp should be updated.
**dp[i] = Math.max(dp[i-2]+nums[i], dp[i-1]);**

Nandha2Tech:
Great Work! Kindly include TC and SC. kindly check house robber problem.
This is how dp should be updated.
**dp[i] = Math.max(dp[i-2]+nums[i], dp[i-1]);**

PR #1471, opened 2023-12-31
done

dewann:


Nandha2Tech (commented):
Great Work! In house robber recursive solution SC would be O(n) (recursive stack space).

Nandha2Tech:
Great Work! In house robber recursive solution SC would be O(n) (recursive stack space).

dewann:
> Great Work! In house robber recursive solution SC would be O(n) (recursive stack space).

Not sure how it will not be 2 pow n.

Nandha2Tech:
TC would be 2 pow n. I was mentioning about SC. It would not be O(1). (because of recursive stack space).

dewann:
> TC would be 2 pow n. I was mentioning about SC. It would not be O(1). (because of recursive stack space).

Sorry my bad.
Got it thanks.

PR #1472, opened 2023-12-31
Completed DP 1

goodguygoddy:


Problem_1.cpp
```diff
@@ -0,0 +1,29 @@
+/*
+ * Time Complexity : O(n*amount)
+ * Space Complexity : O(1)
```
ishika2741: Check space complexity

Problem_2.cpp
```diff
@@ -0,0 +1,64 @@
+/*
+ * Time Complexity : O(n)
+ * Space Complexity : O(1)
```
ishika2741: Check space complexity

ishika2741 (commented):
Good work!

ishika2741:
Good work!

Problem_1.cpp
goodguygoddy: Yes, space complexity would be O(amount)

Problem_2.cpp
goodguygoddy: Yes, space complexity would be O(n)

PR #1518, opened 2024-03-08
house robber

ParvathiRPai:


vyankateshmisar:
Good job Parvathi. Please try to define space and time complexity in terms of size of the input. Instead of O(nums), please write O(n), where n is the length/size of nums.

vyankateshmisar:
Good job Parvathi. Please try to define space and time complexity in terms of size of the input. Instead of O(nums), please write O(n), where n is the length/size of nums.

PR #1745, opened 2025-02-01
completed DP-1

raaulc:


Erlybird:
Good job! Please mentiont the TC and SC at the top of each solution

Erlybird:
Good job! Please mention the TC and SC at the top of each solution

PR #1847, opened 2025-05-29
Initial Commit

rutujakaushikecan:


super30admin:
- CoinChange.java: Well done! Your solution is correct, efficient, and well-structured. The comments and variable names make the code easy to understand. Keep up the good work!
- HouseRobber.java: Your solution is mostly correct, but the use of TreeSet is unnecessary and adds complexity. The TreeSet does not help in solving the problem and should be removed. Focus on keeping the solution simple and efficient. Also, consider adding more comments to explain your thought process.

PR #1869, opened 2025-07-01
Create house_robber.java

rachanamurty:


super30admin:
Strengths:
- The solution is correct and efficiently solves the problem using dynamic programming.
- The use of variables `skip` and `take` effectively captures the two choices at each step (to rob or not to rob the current house).

Areas for Improvement:
- While the code is already excellent, adding a brief explanation of the approach at the beginning of the method could help others understand the logic more quickly.
- The comment "// <-- dontPick case / 0" could be slightly more descriptive to clarify that `skip` represents the maximum amount if the current house is not robbed.

PR #1880, opened 2025-07-09
Done DP-1

rk-peddibhotla:


super30admin:
- Strengths:
  - The approach to solving the problems is logical and follows dynamic programming principles.
- Areas for Improvement:
  - For coinChange.java, consider optimizing the space complexity by using a 1D array instead of a 2D array. This would reduce the space complexity to O(amount).
  - For houseRob.java, the space complexity could also be optimized to O(1) by using variables to store the previous skip and pick values instead of a 2D array.
  - The class names should follow Java naming conventions (e.g., CoinChange instead of coinChange, HouseRob instead of houseRob).

PR #1890, opened 2025-08-02
completed dp-1

DhanyasriG:


super30admin:
Strengths:
- Both solutions are well-structured and follow a logical approach.
- The time and space complexities are optimal for both problems.
- The code is readable and follows good practices.

Areas for Improvement:
- In the CoinChange solution, remove the redundant print statement when dp[m] is 99999. Only print -1 in that case.
- Add more detailed comments explaining the logic, especially for the HouseRobber solution where the logic is not immediately clear from the code alone.
- Ensure that all edge cases are handled properly, such as when n=1 in the HouseRobber solution (which is already handled correctly).

PR #1908, opened 2025-08-28
Implemented DP-1

johar-inderpreet:
Implemented DP-1


super30admin:
- Strengths:
  - The student has a good understanding of dynamic programming and has applied it effectively to solve both problems.
  - The student has provided test cases in the main method to verify the correctness of the solutions.
- Areas for Improvement:
  - For the CoinChange problem, the student could consider initializing the dp array with Integer.MAX_VALUE instead of Integer.MAX_VALUE - 10, and then handle the overflow condition separately if needed. This would make the code more intuitive.
  - For the HouseRobber problem, the student could optimize the space complexity further by using only two variables to store the previous two values instead of an entire array, reducing the space complexity to O(1).

PR #1918, opened 2025-08-31
DP-1 V1

v-hemish:


super30admin:
Strengths:
- Good code structure and readability in both solutions.
- Correct implementation and analysis for the house robber problem (problem2.py).
- Appropriate use of comments to explain the approach.

Areas for Improvement:
- For the coin change problem (problem1.py), the DP table initialization and update logic need to be corrected. The current implementation doesn't properly consider all coins for each amount. You should initialize dp[0] = 0 and all others to INF, then iterate through each coin and update the dp array for all amounts from that coin to the target amount.
- Consider adding more test cases to verify your solutions, especially edge cases like empty coin list or amount = 0.
- For problem2.py, you could optimize space further by using just two variables instead of the entire dp array, since you only need the previous two values at each step.

PR #1920, opened 2025-09-01
Completed DP-1

candida17:


super30admin:
- Problem-1.java: Your solution is correct and efficient. However, consider optimizing the space complexity by using a 1D array instead of a 2D array. This would reduce the space complexity to O(n).
- Problem-2.java: Your solution is correct and efficient. For further optimization, consider using two variables to store the previous two values instead of an array, reducing the space complexity to O(1).
- Both solutions are well-commented and easy to understand. Keep up the good work!

PR #1927, opened 2025-09-29
Add Coin Change and House Robber DP solutions

ravithakur2k:
Introduced CoinChange.py and HouseRobber.py with both top-down (DFS with memoization) and bottom-up (iterative DP) approaches for each problem. These implementations provide efficient solutions to classic dynamic programming problems.


super30admin:
Strengths:
- The student has implemented both recursive (DFS with memoization) and iterative (DP) approaches for both problems, demonstrating a good understanding of different solution strategies.
- The code is well-commented, making it easy to understand the logic and approach.
- The time and space complexity analysis is accurate and clearly stated.

Areas for Improvement:
- In the CoinChange problem, the student could consider using a 1D DP array instead of a 2D array to further optimize space complexity, as the problem can be solved with O(N) space.
- The helper function in the HouseRobber problem could be made more concise by combining the base cases into a single condition if possible.
- The student could add more edge case handling, such as empty input lists, to make the solutions more robust.

PR #1928, opened 2025-10-10
DP-1 solutions

kalyan-ng22:


super30admin:
- CoinChange.java: Great job on implementing the dynamic programming solution correctly. To further optimize space, consider using a 1D array instead of a 2D array. This would reduce the space complexity to O(n) while maintaining the same time complexity.
- HouseRobber.java: Excellent implementation with optimal time and space complexity. The use of variables instead of an array for tabulation is a good optimization. Keep up the good work!

PR #1944, opened 2025-11-20
DP-1

SHETTYSAMARTH30:
Completed DP-1 homework.


super30admin:
- CoinChange.java: The solution is well-implemented and efficient. One minor improvement could be to use a 1D array for the DP table to optimize space further, but the current approach is perfectly valid.
- HouseRobber.java: The solution is excellent. One suggestion is to consider using two variables instead of an array to store the previous two values, which would reduce space complexity to O(1). However, the current approach is clear and correct.

Overall, the student has demonstrated a strong understanding of dynamic programming and has provided efficient and correct solutions. The code quality is high, and the solutions are well-commented.

PR #1952, opened 2025-11-25
Add coin change and house robber solutions

melvinmilkiyas:


super30admin:
- Strengths:
  - Both solutions are correct and efficient, with better time and space complexity than the reference solutions.
  - The code is well-structured and readable, with good use of comments to explain the approach.
  - The dynamic programming approach is implemented correctly in both solutions.

- Areas for Improvement:
  - In the coin change solution, the condition `if dp[cols-1] > amount` could be more accurately written as `if dp[cols-1] == float('inf')` to match the problem's requirements.
  - In the house robber solution, the variable `current` is returned, but it might be clearer to return `prev1` since `current` is updated to `prev1` in the last iteration.

PR #1970, opened 2026-01-06
Done Dp-1

vaishnavi2231:


super30admin:
- Strengths:
  - Both solutions are correct and implement dynamic programming effectively.
  - The time complexities are optimal for both problems.

- Areas for Improvement:
  - For the coinChange problem, consider optimizing the space complexity to O(n) by using a 1D DP array. This would reduce the memory usage without affecting the time complexity.
  - For the rob problem, consider optimizing the space complexity to O(1) by using two variables to keep track of the previous two states instead of an entire array. This would make the solution more memory-efficient.
  - In the coinChange solution, the final check for `dp[i][j]` should use the last indices of the DP table (`dp[len(coins)][amount]`) instead of the loop variables `i` and `j`, which might not be in scope or could lead to incorrect results if not handled carefully.

PR #1977, opened 2026-02-12
Add implementation for leetcode problems 198, 322

rishigoswamy:


super30admin:
Strengths:
- The student has implemented the standard bottom-up dynamic programming solution correctly.
- The time and space complexity are optimal.
- The code is concise and easy to understand.

Areas for improvement:
- The parameter name `amounts` should be changed to `amount` to match the problem statement and avoid confusion.
- The documentation at the top of the file is incorrect: it describes the House Robber problem. This should be fixed to describe the Coin Change problem.
- Adding inline comments to explain the dp array initialization and the inner loop would make the code more readable.
- The return statement can be written as `return dp[amounts] if dp[amounts] != amounts+1 else -1` instead of `dp[-1]` for clarity. Using `dp[-1]` is correct because it is the last element, but explicitly using `dp[amounts]` is more straightforward.

###

PR #1978, opened 2026-02-12
[Using DP to solve rob and coin change]

takchiks:


super30admin:
Strengths:
- You used dynamic programming with a 1D array, which is an efficient approach for the coin change problem.
- You correctly handled the initialization of the dp array with `Integer.MAX_VALUE - 1` to avoid integer overflow.
- Your time and space complexity analysis for coinChange is correct.

Areas for improvement:
- The inner loop in coinChange can be optimized by starting from `j = coins[i]` instead of from 0. This would avoid the condition check `if (j - coins[i] >= 0)` for every j, making the code more efficient and cleaner.
- You included a solution for the "rob" problem in the same file, which is not related to the coin change problem. This might be a mistake. In the future, ensure that your submission only contains the solution for the problem at hand.
- Add comments to explain your approach for the coinChange method. This helps in understanding your code.
- Consider using more descriptive variable names. For example, instead of `i` and `j`, use `coin` and `amt` or similar.

Overall, your solution for coinChange is correct and efficient. However, the unrelated code should be removed.

PR #1981, opened 2026-02-15
Working Implementation

avcode3:


super30admin:
Strengths:
- You implemented the dynamic programming solution correctly, which is the standard approach for this problem.
- Your solution has optimal time complexity (O(m*n)), which is efficient for the given constraints.

Areas for improvement:
- Avoid using magic numbers like 99999. Instead, use a value that is clearly defined, such as `amount + 1` (since the maximum number of coins cannot exceed amount) or `float('inf')`. This makes the code more robust and readable.
- Consider optimizing the space complexity from O(m*n) to O(n) by using a 1D DP array. This would reduce memory usage without affecting time complexity.
- Use more descriptive variable names. For example, instead of `my_arr`, use `dp` or `coin_change_table` to make the code self-documenting.
- Although not necessary, adding comments to explain the base cases and the recurrence relation would be helpful for others reading your code.

For the DP solution with space optimization, you can do:
```
    dp = [amount + 1] * (amount + 1)
    dp[0] = 0
    for coin in coins:
            dp[j] = min(dp[j], 1 + dp[j - coin])
    return dp[amount] if dp[amount] != amount + 1 else -1
```
This approach uses a 1D array and iterates over each coin and then over amounts from the coin value to the target amount.

super30admin:
Strengths:
- You have implemented a dynamic programming solution that is efficient in time (O(m*n)), which is the standard approach for this problem.
- You correctly handled the base cases: initializing the first row (for no coins) to a large number except for amount 0, and the first column to 0.

Areas for improvement:
- Avoid using magic numbers like 99999. Instead, use `float('inf')` or define a constant (e.g., `INF = 10**5`) to represent infinity. This makes the code more robust and readable.
- Consider optimizing the space complexity by using a 1D DP array. Since the current state only depends on the previous row and the current row, you can use a single array of size (amount+1) and update it iteratively.
- Use more descriptive variable names. For example, `dp` instead of `my_arr` would be more conventional for dynamic programming solutions.
- Although not necessary, adding comments to explain the steps can improve readability.

Here is an optimized version of your code with space optimization and better practices:
```
class Solution:
        n = amount
        dp = [float('inf')] * (n + 1)
        dp[0] = 0
        for coin in coins:
            for j in range(coin, n + 1):
                dp[j] = min(dp[j], dp[j - coin] + 1)
        return dp[n] if dp[n] != float('inf') else -1
```
This version uses O(n) space and is more efficient in terms of memory.

PR #1985, opened 2026-02-25
Adding the solutions

subbu4061:


super30admin:
For the coin change problem:
- Solution 1 (DP) is well-implemented. However, the condition for returning -1 is a bit arbitrary. It would be better to define a constant for the initial value (e.g., INF = Integer.MAX_VALUE - 5) and then compare matrix[amount] to INF. Also, note that the inner loop could be optimized by starting from the coin value instead of from 1 and then checking condition. But your current approach is acceptable.

- Solution 2 (Recursion) is a good attempt at a recursive solution, but without memoization it is not efficient for large inputs. You should consider adding memoization to avoid repeated calculations. For example, you could use a 2D memo array with dimensions [coins.length][amount+1] to store the result for state (idx, amount). This would reduce the time complexity to O(coins.length * amount), which is acceptable.

Also, in Solution 2, you subtract 10 from Integer.MAX_VALUE to avoid overflow when adding 1. But it would be better to use a constant for the invalid value.

Overall, you demonstrated understanding of both DP and recursive approaches. However, for the recursive solution, you should always consider memoization to avoid exponential time.

